PR #62, opened 2020-06-26
Added functions for p_chance calculation using eftekhari et al 2017

KshitijAggarwal:
Added functions which use the methods in section 2 of [Eftekhari et al 2017](https://ui.adsabs.harvard.edu/abs/2017ApJ...849..162E/abstract) to calculate the probability of chance association of a galaxy with an FRB. It reads the data used by [Driver et al](https://iopscience.iop.org/article/10.3847/0004-637X/827/2/108#apjaa28a0t2), does a spline fit on it to obtain the relation between r-band magnitude and number counts. These number counts are then used to calculate p_chance as defined in [Bloom et al](https://iopscience.iop.org/article/10.1086/338893/pdf).


frb/galaxies/hosts.py
```diff
@@ -84,3 +88,102 @@ def random_separation(catalog, wcs, npix, trim=1*units.arcmin, ntrial=100):
...
+    return np.max([2*R_frb, np.sqrt(R_0**2 + 4*R_h**2)])
+
+
+def read_r_mags(data_table_path):
+    """
+    Reads data used in Driver et al (2016).
+    https://iopscience.iop.org/article/10.3847/0004-637X/827/2/108
+
+    Args:
+        data_table_path: Path to the fits file with data
```
profxj, line 115: If this file is relatively small in size (e.g. <10Mb), please
add in somewhere sensible under `data/` and then you 
can load it directly from the repository.

frb/galaxies/hosts.py
```diff
@@ -84,3 +88,102 @@ def random_separation(catalog, wcs, npix, trim=1*units.arcmin, ntrial=100):
...
+        array: magnitude bin
+        array: cosmic variance
+
+    """
+    table = fits.open(data_table_path)
+    data = table[1].data
+    r_mask = np.concatenate((np.where(data['Filtername'] == 'r')[0],
+                             np.where(data['Filtername'] == 'F606W')[0]))
+    _magbin = data['MagBinCentre'][r_mask]
+    args = _magbin.argsort()
```
profxj, line 128: maybe use a different variable name than `args` which has special meaning in Python.

frb/galaxies/hosts.py
```diff
@@ -84,3 +88,102 @@ def random_separation(catalog, wcs, npix, trim=1*units.arcmin, ntrial=100):
...
+
+
+def prob_eb17(R_frb, m, R_0=0.2, R_h=0.25, ret_numgal=False):
+    """
+    Calculates chance association probability of a galaxy to an FRB
+    Taken from: 
+        Section 2 of https://ui.adsabs.harvard.edu/abs/2017ApJ...849..162E/abstract
+
+    Args:
+        R_frb: The 1 sigma localization radius of the FRB
```
profxj, line 160: please define the types of each arg

frb/galaxies/hosts.py
profxj: Ah, I see you do so below.  Ignore this then.  :)

frb/galaxies/hosts.py
```diff
@@ -84,3 +88,102 @@ def random_separation(catalog, wcs, npix, trim=1*units.arcmin, ntrial=100):
...
+    
+    Args:
+        R_frb: The 1 sigma localization radius of the FRB
+        R_0: Radial angular separation between the FRB position and a presumed host
+        R_h: Galaxy half light radius
+    
+    Returns:
+        float: radius (in arcseconds)
+    """
+    return np.max([2*R_frb, np.sqrt(R_0**2 + 4*R_h**2)])
```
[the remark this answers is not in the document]
KshitijAggarwal: `2*R_h` is always less than `np.sqrt(R_0**2 + 4*R_h**2)`, so it would never be used as we are taking a `max`.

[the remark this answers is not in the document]
profxj: ah, right!

PR #83, opened 2020-11-16
Hst galfit

SunilSimha:
Modifications to the GALFIT wrapper to work on HST images. Include the ability to parse two-component sersic fits.


profxj (changes requested):
Looks good.  Does `defs.py` in `galaxies/` need updating?

How about the `parse_galfit()` call?

Maybe add a test in `test_galfit.py` ?

SunilSimha:
> Looks good. Does `defs.py` in `galaxies/` need updating?

I don't think so. The parameter names haven't changed.

> How about the `parse_galfit()` call?

Now updated to correctly read two components.

> Maybe add a test in `test_galfit.py` ?

Added.

profxj (approved):
All good!

PR #103, opened 2021-03-23
Cigale hotfix

SunilSimha:
Two minor bugs related to photometry and cigale fixed.


profxj (changes requested):
Thanks for this.  Can we add a test which uses this code
and reports an answer we know is truth?  That way we'll
catch it if future changes manage to break it.

profxj (approved):
All good!

PR #117, opened 2021-08-30
no longer ignores negative sign on floats

SunilSimha:
@Lachimax, here is the bug fix for the missing negative signs on the floats. 

@profxj, I've updated a regex string that matches floats in the `parse_galfit` function. Until now, the regex matching was ignoring negative signs when looking for floats in the galfit output file. I've gotten away with it as it only affected the PA until now and I don't think we had used that information much. Nevertheless, it might be prudent to re-run `parse_galfit` on all galfit outputs we've generated since Mannings+2020.


Lachimax:
Has fixed my issue, thank you!

Lachimax (commented):
Works for me now!

profxj (approved):
thx for this.

And I will re-run all galfit on the `refactor`
branch before we merge it.

PR #133, opened 2022-02-18
Fix output to file of the DM-redshift grid on first run

fjankowsk:
The commit fixes a missing argument that is necessary for the `frb_pz_dm` script, in particular for the generation of the DM-redshift grid on the first run. No pre-generated grid file is currently provided.


profxj (approved):
thx for this!

PR #140, opened 2022-07-22
The - sign is incorrect.

SunilSimha:
As the title suggests. Adding the - sign cancels the 90 degree rotation.


frb/figures/finder.py
```diff
@@ -208,8 +208,8 @@ def generate(image, wcs, title, flip_ra=False, flip_dec=False,
         
         pa_deg = pa.to('deg').value
         # according to SkyRectangularAperture the theta is for the "width", thus we should conform that geometry here.
-        # for PA = 0 we want the slit to be oriented North-South, thus we need to rotate by +90 degrees and apply - sign 
-        aper = SkyRectangularAperture(positions=slit_coords, w=width, h=length, theta=-1*(pa+90*units.deg))  # PA should increase from North to East 
+        # for PA = 0 we want the slit to be oriented North-South, thus we need to rotate by +90 degrees
+        aper = SkyRectangularAperture(positions=slit_coords, w=width, h=length, theta=(pa+90*units.deg))  # PA should increase from North to East
```
ntejos, line 213: The documentation of `SkyRectangularAperture` says: ```The position angle (in angular units) of the rectangle “width” side. For a right-handed world coordinate system, the position angle increases counterclockwise from North (PA=0)```. This the sign will depend on wether we are in a right-handed versus left-handed WCS. Is there a way to know this from the `wcs` object? If so, perhaps and if statement to cover both possible cases is needed to sort the sign issue.

ntejos (approved):
Left a comment on the PA sign issue

frb/figures/finder.py
SunilSimha: Now taken care of. Thanks!

PR #141, opened 2022-07-27
Corrections to the likelihood function used in dm/mcmc.py

JoschaJ:
This fixes an error in the normalization of the likelihood function "all_prob" in frb/dm/mcmc.py. The issue was that the normalization integral goes over Delta instead of DM and therefore requires a factor <DM_cosmic> that comes from the substitution dDM = dDelta * <DM_cosmic>. The change significantly affects the posterior distribution of the MCMC.

The function "all_prob" has also been rewritten for a significant speed-up. It is replaced by "log_likelihood_variable_step". The  similar function "log_likelihood" is usually slower, but more robust. It is left, but currently unused.


profxj (approved):
Thx for this!  

I am approving without a careful review
mainly because we have transitioned away
from this analysis to using that in:

https://github.com/FRBs/zdm

PR #142, opened 2022-09-02
Renaming to avoid MacOS file name issues

SunilSimha:
As described above. I have the filters everywhere I could find them. I imagine this will mean some legacy code to handle SOAR photometry might break but that's a problem for Future Sunil to handle.

For now, in case, you see a potential failure mode that I've missed, please let me know.


profxj (approved):
Thx for this.

Am approving, but remove `tst.png` in `tests/files` unless it really is needed.

almannin (approved):
thank you!!

PR #145, opened 2022-09-29
implementing get_DMcosmic_from_z() that provides different percentile…

ntejos:
…s on the zdm relation.


frb/dm/prob_dmz.py
```diff
@@ -169,4 +173,67 @@ def grab_repo_grid():
...
+    -------
+    zarray : np.array
+        z values where the DMcosmic was computed
+    DMcosmic_array : np.array
+        DMcosmic values corresponding to the z_array (in a given percentile)
+    
+    """
+    
+    #check z input
+    if isinstance(zarray, float):
```
SunilSimha, line 215: I'd replace this isinstance with `np.isscalar`. This takes care `ints` as well.

frb/dm/prob_dmz.py
```diff
@@ -169,4 +173,67 @@ def grab_repo_grid():
...
+        z_new = np.array([0,zarray])
+    else:
+        z_new = np.array(zarray)
+   
+    # Get the DMcosmic-z grid
+    # This is time consuming, could be done more efficiently
+    print("Calculating the DMcosmic-z grid, this may take a while...")
+    zeval , DM_cosmics, PDF_grid = grid_P_DMcosmic_z(zvals=z_new, beta=beta, F=F, cosmo=cosmo, DM_cosmics=DMevals)
```
SunilSimha, line 223: Now that I think about it, maybe it would be a good idea to just store the default grid of values in a compressed `npz` file. For a 1000x1000 grid of floats, the size should be less than 1 MB. Users wanting to generate custom grids can do so at their leisure.

Alternatively, maybe provide an option to read the grid from a file.

SunilSimha (approved):

PR #151, opened 2023-03-08
DELVE

SunilSimha:
This PR includes
* Renaming of SDSS filter files.
* Adding DELVE filters to the `defs` and the `cigale` wrapper. They're just DECam filters. Also adding 6dF filters.
* A `DELVE_Survey` class for pulling in data from NOIRLAB.


profxj (changes requested):
Great stuff.

Just consider my one question and I'll approve.

profxj (approved):

PR #153, opened 2023-04-04
image access bug fixes

SunilSimha:
I've fixed some of the image access bugs for the `dlsurvey` class and tweaked the code to return the deepest image, i.e. the one with the largest zeropoint as opposed to largest exptime (which is not necessarily unique to a file).


profxj (approved):
looks great, thanks!

PR #181, opened 2024-05-09
DECaLS dr8->dr10

SunilSimha:
Migrating from DR8 to DR10 for our DECaLS data. Might potentially break the CHIME_FFFF_PZ PATH code and hence, I'm requesting that @bandersen441 and @lordrick94 take a look.

Most important change is the way stars are defined. Instead of using the `gaia_pointsource` flag which is no longer available in the `tractor` table, I am using the morphological `type` classification. Stars are now `type=='PSF'` in this new system.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -19,8 +19,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


profxj (approved):
thx for this.

and, yes, we'll need to update `chime_ffff_pz`

maybe you can do so directly @SunilSimha ?

PR #190, opened 2024-10-29
Pkg resources

SunilSimha:
pkg_resources has been deprecated for a long time and in Python 3.12, it is finally non-functional.

This PR Replaces all instances of pkg_resources.resource_filename with importlib_resources.files. The latter returns a PosixPath object instead of as str and so one has to be careful with functions that expect strings in path inputs. All unit tests run without errors pertaining to path names.

There might still be errors because of your external packages which use pkg_resources. This only deals with bugs within the repo.


profxj (approved):
whoa, very helpful.

thanks!

PR #192, opened 2024-10-30
Ps1 strm photoz

SunilSimha:
Downloads PS1 photometric redshifts from Mast Casjobs and merges it with the main photometry table. Mast Casjobs requires its own package to be installed and its import is hidden within the `get_catalog` function, making this an optional dependency.

Further requires the user to create an account and store login credentials as environment variables.


tarraneh (approved):
Works great, thanks!

PR #196, opened 2025-01-22
Setting photoz argument default to True

tarraneh:
Automatically pulls photo-zs for PS1 and adds them to the PATH candidates table. 

Note that this **REQUIRES** a MAST user account: https://mastweb.stsci.edu/mcasjobs/CreateAccount.aspx
And for MAST_CASJOBS_USER and MAST_CASJOBS_PWD environment variables to be set.


SunilSimha (changes requested):
In line 155's `raise` clause, I'd add an additional sentence for those who don't want to draw photoz's. e.g. "You need to ... get your credentials. Or set `photoz=False` in `get_catalog`."

SunilSimha (approved):
Approved

PR #201, opened 2025-03-20
Cigale fixes

SunilSimha:
This PR is to update the CIGALE wrapper and to make it compatible with the code in the master branch of CIGALE (v. 2025.0) . Key changes: 
1. Photometry filter names are updated.
2. Module parameters are now compatible with the latest version.
3. Warning and error messages using the same API as CIGALE.
4. Removes the  `compare_obs_model` code chunk as CIGALE is already generating those data. No need for another output table.

### DIFF
--- a/frb/data/analysis/CIGALE/DEIMOS_r.dat
+++ b/frb/data/analysis/CIGALE/DEIMOS_r.dat
@@ -1,4 +1,5 @@
 #DEIMOS_r
+#photon
 #DEIMOS r band https://www2.keck.hawaii.edu/inst/deimos/filter_list.html
 5500.0 0.0
 5500.3 0.000159287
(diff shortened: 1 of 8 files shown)


muethela (approved):
All looks reasonable to me

frb/data/analysis/CIGALE/DEIMOS_r.dat
```diff
@@ -1,4 +1,5 @@
 #DEIMOS_r
+#photon
```
profxj, line 2: does this need to be done for every file in the folder?

profxj (approved):
thx!

frb/data/analysis/CIGALE/DEIMOS_r.dat
SunilSimha: I checked the other files and the DEIMOS filters were the only ones lacking this metadata.

PR #202, opened 2025-03-21
PS1 tiling

SunilSimha:
A small function that tiles multiple PS1 cone searches to overcome the cone search radius limit imposed by MAST (30'). Useful for planning wide-field surveys based on PS1 targets.


bandersen441 (approved):
Looks good!

PR #203, opened 2025-04-02
KCWI datacube handling

SunilSimha:
This code deals with KCWI datacubes for: 
1. Extracting spectra.
2. Extracting images (including synthetic images with a filter curve).
3. Detecting sources in images and extracting their spectra from the cube.

Refactored some code to make it compatible with the latest `spectral-cube` repository. Also, all relevant functions are now packaged into a convenient class that can be instantiated with data from a reduced cube from PypeIt or KCWI DRP. There's also an updated notebook to showcase the usage of this module.


frb/surveys/survey_utils.py
```diff
@@ -82,6 +83,8 @@ def load_survey_by_name(name, coord, radius, **kwargs):
         survey = NEDLVS(coord, radius, **kwargs)
     elif name == 'TullyGroupCat':
         survey = TullyGroupCat(coord, radius, **kwargs)
+    elif name == 'DESI':
+        survey = DESI_Survey(coord, radius, **kwargs)
```
leungcalvin, line 87: 'DESIY1'?

leungcalvin (changes requested):
Mostly looks good - see my comment about DESI Y1 vs Y3

frb/analysis/kcwi.py
```diff
@@ -113,358 +66,473 @@ def silence_warnings(warncategory):
...
     """
-    assert len(mask_1d) == len(cube.spectral_axis), "Mask length ({:d}) doesn't match cube's spectral axis length ({:d}).".format(len(mask_1d), len(cube.spectral_axis))
-
-    assert mask_1d.dtype == bool, "Mask must be a boolean array."
 
-    mm = _spectral_tile(mask_1d, cube)
-    return cube.with_mask(mm)
+    # Clean up the spectra so that the region of overlap is
+    # set to NaN.
+    delta_lambda_blue = np.median(np.diff(blue_wave))
```
leungcalvin, line 186: nanmedian perhaps better

frb/analysis/kcwi.py
```diff
@@ -113,358 +66,473 @@ def silence_warnings(warncategory):
...
+        """
+        Helper function that tiles 1D array of
+        size cube.spectral_axis.shape and tiles it
+        to the same shape as the cube.
+        Args:
+            array (ndarray): the 1D array that needs to be tiled.
+        Returns:
+            tiled_array (ndarray): an array with the 1D array tiled
+                in the spatial dimension. This has the same shape
+                as the cube.
```
leungcalvin, line 229: Explicitly say shape is `(n_wave, nx, ny)`

frb/analysis/kcwi.py
```diff
@@ -113,358 +66,473 @@ def silence_warnings(warncategory):
...
+
+    def spec_from_ellipse(self,
+                        x0 = 0., y0 = 0., a = 1.,
+                        b = 1., theta = 0.):
+        
+        """
+        Get the spectrum within an elliptical region
+        Args:
+            cube (Spectral Cube): A datacube object
+            x0, y0 (float, optional): Centroid of ellipse
```
leungcalvin, line 362: Specify units, or pixels, if in pixels

frb/analysis/kcwi.py
```diff
@@ -113,358 +66,473 @@ def silence_warnings(warncategory):
...
+        #TODO: Use photutils aperture object for this.
+        # Create 2D mask first
+        ellipse = EllipticalAperture((x0, y0), a, b, theta)
+        # Create a mask from the ellipse
+        aper_mask = ellipse.to_mask(method='center').to_image(self.cube.shape[1:]).astype(bool)
+        
+        return self.spec_from_spatial_mask(aper_mask)
+
+
+    def _make_marz(self, speclist, varspeclist, objects,  wave=None, marzfile="marzfile.fits", tovac = True):
```
leungcalvin, line 380: why is varspeclist not calculated internally from self.varcube?

frb/analysis/kcwi.py
```diff
@@ -113,358 +66,473 @@ def silence_warnings(warncategory):
...
+        # TODO: Actually compute sky background
+        nobjs = len(objects)
+        wcsinfo = self.cube.wcs.celestial
+
+        sky = np.zeros_like(speclist)
+
+        # Convert wavelengths from air to vacuum
+        if wave is None:
+            wave = self.cube.spectral_axis.value
+        if tovac:
```
leungcalvin, line 410: do you need to save the `tovac` keyword to the output file such that you don't lose track of which marz files have `tovac` vs not?

frb/surveys/desi.py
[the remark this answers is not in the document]
SunilSimha: Generally, we just update the code to pull from the latest table. In this case, since DESI data is being hosted by NOIRLAB, it will mean we just change which table we're pulling from. Unless there's a strong need to pull both Y1 and Y3 data, we'll just download the latest at the time.

frb/surveys/dlsurvey.py
```diff
@@ -63,15 +63,17 @@ def _parse_cat_band(self,band):
 
         return table_cols, col_vals, band
 
-    def _gen_cat_query(self,query_fields=None, qtype='main'):
+    def _gen_cat_query(self,query_fields=None, qtype='main', ra_col=None, dec_col=None):
```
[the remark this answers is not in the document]
SunilSimha: This is because the default RA and DEC columns for DESI are not called `ra` and `dec` unlike every other table in NOIRLAB. I'm adding this accommodation here to still be able to use this super class without breaking all of our other survey classes relying on it.

frb/surveys/survey_utils.py
```diff
@@ -34,7 +35,7 @@
 def load_survey_by_name(name, coord, radius, **kwargs):
     """
     Load up a Survey class object for the named survey
-    allowed_surveys = ['SDSS', 'DES', 'NVSS', 'FIRST', 'WENSS', 'DECaL', 
+    allowed_surveys = ['SDSS', 'DES', 'DESI', 'NVSS', 'FIRST', 'WENSS', 'DECaL',
```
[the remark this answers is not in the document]
SunilSimha: No, we just update to the latest each time a new release is out. This will generally involve minimal effort (i.e. changing the table we're pulling from and perhaps renamig a few columns).

frb/analysis/kcwi.py
SunilSimha: It is. The specrtra and varspecs are being produced by spec from spatial mask. This function only takes a list of those calculated values and formats them into a MARZ file.

frb/analysis/kcwi.py
SunilSimha: Good point. I need to look into MARZ docs if there's a consistent way to do this.

frb/analysis/kcwi.py
SunilSimha: Done.

frb/analysis/kcwi.py
SunilSimha: Done.

frb/analysis/kcwi.py
SunilSimha: So MARZ just looks for a header keyword. I've now switched to providing that instead of doing the conversion myself. Helps keep track explicitly and prevents multiple conversions.

profxj (approved):
thanks for all of this!

PR #205, opened 2025-04-22
pzdm plotting 

lluism:
just small touches to the plots of p(z|DM) and magnitudes

### DIFF
--- a/frb/galaxies/mag_dm.py
+++ b/frb/galaxies/mag_dm.py
@@ -1,6 +1,7 @@
 """ Estimate magnitude range for a host FRB given DM and plot it """
 
 import numpy as np
+import os
 
 #import healpy as hp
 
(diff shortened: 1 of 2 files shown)


frb/galaxies/mag_dm.py
```diff
@@ -77,11 +78,7 @@ def r_vs_dm_figure(z_min, z_max, z, PzDM, outfile='fig_r_vs_z.png',
         ax.plot(good_hosts.z, good_hosts.m_r, 'ok',
                 label='Secure hosts')
 
-    # plot
-    # L curves
-
-
-    zvals = np.linspace(0.021, z_max+0.5, 200)
+    zvals = np.linspace(0.021, min(z_max+0.5,4.), 200)
```
profxj, line 81: do you want the user to be able to input 
  an array of `zvals` as an option?

profxj (approved):
one minor thing to consider

frb/galaxies/mag_dm.py
lluism: great suggestion. I added that.

PR #206, opened 2025-04-23
Survey hotfixes

SunilSimha:
Merging catalogs no longer fails, especially with Numpy >2.0. Also, PS1 tiling had a bug that's now fixed.


bizarreboa (approved):
Works! Had to make small changes in the HostPipeline code though, GALEX would find 2 objects and code was unhappy. Will push those changes in chime-ffff-pz

PR #207, opened 2025-04-24
Catalog searches

SunilSimha:
Searching cluster catalogs in Vizier.


leungcalvin:
almost works, but I had to change all the `super(XX,self).__init__` -> `VizierCatalogSearch.__init__` and then it seemed to work. 
image

leungcalvin:
This is a really cool tool. Call me nosy, but I was playing with this a bit today: I guess `clean_catalog` is not fully mature yet, but I'm not getting any hits with a 3 degree radius? Seems weird.

```
import importlib
from frb.surveys import cluster_search
importlib.reload(cluster_search)
import astropy.coordinates as ac
import astropy.units as u

hg = ac.SkyCoord(ra = '01h45m52.26s', dec = '35d06m28.19s')

for survey in [cluster_search.TullyGroupCat,cluster_search.WenGroupCat,cluster_search.UPClusterSZCat,cluster_search.TempelClusterCat,cluster_search.ROSATXClusterCat]:
    a = survey(coord = hg, radius = 3 * u.deg)
    a._get_catalog()
    print(a.catalog)
```
image

SunilSimha:
@leungcalvin you need to use `get_catalog()` instead of `_get_catalog()`. The latter is a helper function from the parent class.

PR #208, opened 2025-05-22
M31 refactor

lordrick94:
Adding logMCGM input for M31 mNFW object

### DIFF
--- a/frb/halos/models.py
+++ b/frb/halos/models.py
@@ -1,7 +1,5 @@
 """ Module for DM Halo calculations
 """
-from __future__ import print_function, absolute_import, division, unicode_literals
-
 import numpy as np
 import pdb
 import warnings
@@ -544,7 +542,7 @@ class ModifiedNFW(object):
 
     Parameters:
         log_Mhalo: float, optional
-          log10 of the Halo mass (solar masses)
+          log10 of the Halo mass [dark matter + baryons] (solar masses)
         c: float, optional
           concentration of the halo
         f_hot: float, optional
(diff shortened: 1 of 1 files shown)


frb/halos/models.py
```diff
@@ -591,16 +603,16 @@ def __init__(self, log_Mhalo=12.2, c=7.67, f_hot=0.75, alpha=0.,
...
-        if cosmo is None:
-            self.rhoc = 9.2e-30 * units.g / units.cm**3
-            self.fb = 0.16       # Baryon fraction
-            self.H0 = 70. *units.km/units.s/ units.Mpc
-        else:
-            self.rhoc = self.cosmo.critical_density(self.z)
-            self.fb = cosmo.Ob0/cosmo.Om0
-            self.H0 = cosmo.H0
+        self.rhoc = self.cosmo.critical_density(self.z)
+        self.fb = cosmo.Ob0/cosmo.Om0
```
SunilSimha, line 613: I suggest `fb ` be defined within `__init__()`. I see lines 586 and 588 are using this definition already, so it makes sense to just reference `self.fb` for both simplicity of reading and to ensure changes are propagated across the board in the future.

frb/halos/models.py
```diff
@@ -1088,7 +1100,9 @@ class M31(ModifiedNFW):
     Taking mass from van der Marel 2012
 
     """
-    def __init__(self, log_Mhalo=12.18, c=7.67, f_hot=0.75, alpha=2, y0=2, **kwargs):
+    def __init__(self, 
+                 log_Mhalo=12.18, c=7.67, 
+                 alpha=2, y0=2, **kwargs):
```
SunilSimha, line 1105: `f_hot` is not an explicit parameter anymore? Intentional?

frb/halos/models.py
```diff
@@ -544,12 +542,15 @@ class ModifiedNFW(object):
...
     Parameters:
         log_Mhalo: float, optional
-          log10 of the Halo mass (solar masses)
+          log10 of the Halo mass [dark matter + baryons] (solar masses)
         c: float, optional
           concentration of the halo
         f_hot: float, optional
           Fraction of the baryons in this hot phase
           Will likely use this for all diffuse gas
+          This is exactly the mass of the CGM divided by the
```
SunilSimha, line 551: ... mass of the CGM within r200...

SunilSimha (changes requested):
Requesting minor edits. Thanks!

frb/halos/models.py
lordrick94: Fixed

frb/halos/models.py
lordrick94: Done

profxj (approved):
good stuff

do we need a new test?

SunilSimha (approved):
Looks good now. The M31 halo unit test seems to be passing at the moment but as X said, if you think we should add an assert statement or two to test the new definitions, @lordrick94 we should.

PR #232, opened 2026-03-11
Rerun CRAFT + PATH

profxj:
New coords for a number of CRAFT FRBs which led to more PATH
runnin


cwjames1983 (commented):
All good, thanks @profxj !

cwjames1983 (approved):

PR #235, opened 2026-05-05
Euclid Survey

SunilSimha:
In light of the discussion had at the Host WG, this PR adds the Euclid Survey class to the FRB repo. This can only query public data (i.e., Q1 outputs) and will not be a replacement for what our French colleagues are doing. Still, all data will eventually become public, and at that time, one will likely need to make only minor changes to this code.

In addition to the usual catalog search and cutout query, this adds a `get_spectrum` feature, which can download Euclid spectra for a single source if available.

### DIFF
--- a/docs/api/surveys.euclid.rst
+++ b/docs/api/surveys.euclid.rst
@@ -1,0 +1,7 @@
+frb.surveys.euclid
+==================
+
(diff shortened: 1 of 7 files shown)


bizarreboa (approved):
tested the test_euclid test and it works so I'm sure it's good!

PR #236, opened 2026-05-05
some small fixes and edits to work with modap

bizarreboa:
mostly added a way to change the cigale SED plot to be a png so that text can be added to it, and then several various fixes for small errors


frb/galaxies/cigale.py
```diff
@@ -452,3 +469,54 @@ def host_run(host, cut_photom=None, cigale_file=None):
         sfh_file = cigale_file.replace('CIGALE', 'CIGALE_SFH')
         os.system('mv {:s}/{:s}_SFH.fits {:s}'.format(host.name, host.name, sfh_file))
     return
+
+def add_text_SED(host, cigale_results, out_name=None):
```
SunilSimha, line 473: Seems a little hacky so are you certain there's no internal CIGALE plot argument that does this already? If not, feel free to just resolve this comment.

frb/galaxies/photom.py
```diff
@@ -187,9 +187,13 @@ def extinction_correction(filt, EBV, RV=3.1, max_wave=None, required=True):
     
     source_flux = 1.
     #calculate linear correction
-    delta = np.trapezoid(throughput * source_flux * 
-                     10 ** (-0.4 * Alambda), x=wave) / np.trapezoid(
-                         throughput * source_flux, x=wave)
+    if getattr(np, "trapz", None) is not None:
```
SunilSimha, line 190: Maybe worth adding a deprecation warning here. I believe this is for Numpy < 2.0, and so I don't see a reason to support this long term.

SunilSimha: Alternatively, switching over to the scipy implementation if the naming there is a little more stable.

frb/galaxies/photom.py
```diff
@@ -250,12 +254,13 @@ def correct_photom_table(photom, EBV, name, max_wave=None, required=True):
...
         except:
             embed(header='187 in photom')
-        # SDSS
-        if 'SDSS' in filt:
-            if 'extinction_{}'.format(filt[-1]) in photom.keys():
-                print("Appying SDSS-provided extinction correction")
-                cut_photom[key] -= cut_photom['extinction_{}'.format(filt[-1])]
-                continue
+        print('Correcting filter {} for Galactic extinction'.format(filt))
+        # SDSS ### removing this because SDSS correction is sometimes different for very dusty sightlines, but TBD
```
SunilSimha, line 258: I would trust SDSS extinction estimates more because I imagine they got their E(B-V) values at the location where they took their spectra as opposed to our method of using the nearest measurement within a few degrees. Worth looking into why they are different.

frb/scripts/r_vs_dm.py
```diff
@@ -165,4 +165,4 @@ def r_vs_dm(outfile='fig_r_vs_z.png',
 ''' Run me '''
 # Command line execution
 if __name__ == '__main__':
-    r_vs_dm()
\ No newline at end of file
+    r_vs_dm()
```
SunilSimha, line 168: Can you check that there's a corresponding entry in the new `pyproject.toml` to make sure this script is installable when `pip install`ing the repo?

frb/surveys/catalog_utils.py
```diff
@@ -100,7 +100,11 @@ def match_ids(IDs, match_IDs, require_in_match=True):
     """
     rows = -1 * np.ones_like(IDs).astype(int)
     # Find which IDs are in match_IDs
-    in_match = np.in1d(IDs, match_IDs)
+    try:
+        in_match = np.isin(IDs, match_IDs)
+    except AttributeError:
+        # Older NumPy versions
```
SunilSimha, line 106: I was under the impression that `np.isin` has always been available, even in v 1.26. If so, maybe we can get rid of this exception handling?

frb/surveys/delve.py
```diff
@@ -20,14 +20,19 @@
 photom = {}
 photom['DELVE'] = {}
 photom['DELVE']['DELVE_ID'] = 'quick_object_id'
+# photom['DELVE']['DELVE_ID'] = 'coadd_object_id'
```
SunilSimha, line 23: Seems like these comments are not changing the original functionality or documenting anything. I suggest getting rid of the comment blocks in this file.

frb/surveys/galex.py
```diff
@@ -63,7 +63,7 @@ def get_catalog(self,query_fields=None, print_query=False):
             query_fields = _DEFAULT_query_fields
         else:
             query_fields = _DEFAULT_query_fields+query_fields
-        
+        import astropy.units as u
```
SunilSimha, line 66: Is this import being used anywhere in the file? If not, please remove.

frb/surveys/survey_utils.py
```diff
@@ -270,13 +270,64 @@ def search_all_surveys(coord:SkyCoord, radius:u.Quantity, include_radio:bool=Fal
...
+    """
+    if len(cat) == 0:
+        return cat
+
+    if mag_cols is None:
+        from frb.galaxies.defs import valid_filters
+        mag_cols = valid_filters
+
+    # require separation column already computed in arcmin/arcsec/whatever
+    sep = cat['separation']
```
SunilSimha, line 309: This is not guaranteed to be present for some surveys, I think.

I really need to make the survey catalogs and method inputs uniform. 😬

frb/surveys/survey_utils.py
```diff
@@ -270,13 +270,64 @@ def search_all_surveys(coord:SkyCoord, radius:u.Quantity, include_radio:bool=Fal
...
+            to consider. If None, no maximum separation
+            is applied.
+
+    Returns:
+        best_row (Table): A 1-row table with the best row.
+    """
+    if len(cat) == 0:
+        return cat
+
+    if mag_cols is None:
```
SunilSimha, line 304: Pretty sure there's a `catalog_utils` function called `get_mag_cols` or some such that just gives you the columns that are present in a table. This is being used for extenction correction I believe. Prevents you from running the if statement in the loop over columns below.

SunilSimha (changes requested):
Requesting changes.

For whatever reason, GitHub is not allowing me to create more comments. So here are the remaining ones:

Line 314 of `survey_utils`

I'd make the table `masked` here and add `good` masks to individual columns. Then, it becomes easy to get the combined mask by getting the "OR" of the individual column masks. Also preserves more info if the user wishes for more granularity later.

e.g.

```python
cat = Table(cat, masked=True)
has_good = np.zeros_like(tab, dtype=bool)

for col in mag_cols:
    tab[col].mask = np.isfinite(tab[col]) & tab[col]< 90 & (tab > 0)
    has_good |= tab[col].mask 
```

Line 273 of `survey_utils`

This doesn't make sense here. The combined cat for a large cone search should return several rows. Therefore, picking the best row should be a separate function implemented elsewhere. If MODaP needs this, then MODaP must implement it on whatever this function returns.

frb/galaxies/cigale.py
bizarreboa: Checked — `pcigale-plots sed` has no text-annotation option; the labels are baked
into the matplotlib figure it writes, so there's nothing to hook into. Keeping the
helper.

frb/galaxies/photom.py
bizarreboa: Went with `np.trapezoid`

frb/galaxies/photom.py
bizarreboa: Agreed, the SDSS-provided value is preferred again, now behind a
`_valid_extinction_value()` guard so masked / NaN / -999 entries fall through to the
Gordon 2024 curve instead of silently corrupting the magnitude. Also added an optional
`ext_methods` dict so callers can record which correction was used per filter.

frb/scripts/r_vs_dm.py
bizarreboa: Checked — there's no entry, and I don't think it needs one: `r_vs_dm.py` has no
`cli()`/`main()`, and it's one of five scripts (`iteration`, `pz_given_dm`,
`pzdm_mag_crossmatch`, `r_vs_dm`, `random_assoc`) run directly rather than installed.

frb/surveys/catalog_utils.py
bizarreboa: Gotcha, dropped the try/except.

frb/surveys/delve.py
bizarreboa: Removed all the commented-out blocks I added.

frb/surveys/galex.py
bizarreboa: It is used — `self.radius.to(u.deg)` a few lines down — and the module has no top-level
units import, so removing it outright breaks `get_catalog()`. Moved it up into the
import block instead.

frb/surveys/survey_utils.py
bizarreboa: Ah, for now added an early return with a warning when `separation` isn't present. Also removed the call from `search_all_surveys`, so this function is now only reached when a caller opts in.

frb/surveys/survey_utils.py
bizarreboa: `catalog_utils._detect_mag_cols` now used instead of iterating
`valid_filters` with a membership test inside the loop (dfa99dd).

bizarreboa:
**On `pick_best_row_by_phot` being called inside `search_all_surveys`:** `test_frbsurveys.test_search_all` asserts
`len(combined_cat) == 2`, so collapsing the result to one row broke it — it just never
ran, being `@remote_data`. Removed the call in 8f5922f; `search_all_surveys` returns
every row in the cone again, and `pick_best_row_by_phot` stays as a public helper.

**On the masked-table suggestion:** adopted with one change. The provided snippet
assigns the *good* condition to `.mask` and then ORs the masks together but that would hide the usable photometry, so I inverted the polarity. The returned table is masked, so callers can see which filters actually carried the row.

Also picked up `_detect_mag_cols` and the missing-`separation` guard from the inline
comments. Branch is pushed and ready for another look.

PR #237, opened 2026-05-14
Revamp survey functionality and improve error handling

SunilSimha:
This PR revises the functionality of `frb.surveys` to make it more uniform and adds extensive documentation.

1. All survey classes now try to ensure similar input and output formats. This includes:
a. Ensuring that arguments for the same function are named similarly. e.g. All surveys now use `band` to denote the imaging filter instead of `filt` and `band` being used inconsistently in catalog/image searches.
b. All surveys now define `get_image` method for downloading FITS images explicitly. `get_cutout` is explicitly for downloading PNG/JPEG fileds. To ensure a smooth transition to this new paradigm, surveys that previously used `get_cutout` to download FITS images (e.g. `DECaL_Survey`) provide a `DeprecationWarning` and continue allowing FITS downloads. This will be removed in subsequent releases of the software.
c. Empty tables from catalog searches yielding no rows are handled uniformly. All empty tables should now return a 0-length table with standard column names that are being queried by default. Previously, some surveys would return `None` or an empty table with some named columns or just an empty table with nothing in it. 
d. All bad photometry (0 > mag or mag > 30) are now being masked with a value of `-99.0`. Corresponding errors are also masked in addition to bad errors (0 > err or err > 5).
2. Unit tests in `test_frbsurveys` are renewed to test all the abovementioned functionality.
3. New`.rst` files document changes to the API. In addition, there is the `docs/nb/Surveys.ipynb` notebook that shows a user how to query every supported survey for catalogs, images, and cutouts (where supported). 

Also included are bug fixes:
1. No more `np.in1d`. Replaced by `np.isin`.
2. VISTA image searches directly download from VSA as opposed to the now-defunct Astro Datalab Link.
3. Surveys that lacked FITS image download support, such as GALEX, SDSS, and 2MASS are now supported via the `SkyView` API provided by `astroquery`.


frb/surveys/catalog_utils.py
```diff
@@ -25,15 +25,102 @@ def clean_heasarc(catalog):
         catalog[key].unit = units.deg
 
 
-def clean_cat(catalog, pdict, fill_mask=None):
+def _is_numeric_column(column):
```
profxj, line 28: here and throughout, please make sure the args and returned items are all in the doc strings

frb/surveys/catalog_utils.py
```diff
@@ -379,12 +592,6 @@ def convert_mags_to_flux(photometry_table, fluxunits='mJy', exact_mag_err=False)
         fluxtable[err][baderr] = flux_err[baderr]
         fluxtable[err][~baderr] = flux_err[~baderr]*convert
 
-        # Upper limits -- Record as 3sigma
-        #   and set error to -99.
-        uplimit = photometry_table[err] == 999.
```
profxj, line 384: I worry these 999. mattered

Are we sure they didn't?

frb/surveys/catalog_utils.py
```diff
@@ -476,7 +683,7 @@ def xmatch_and_merge_cats(tab1:Table, tab2:Table, tol:units.Quantity=1*units.arc
     if (len(not_matched_tab1)!=0)&(len(not_matched_tab2)!=0):
         outer_join = join(not_matched_tab1, not_matched_tab2,
                     keys=['ra','dec'], join_type='outer', table_names=table_names)
-        merged = vstack([inner_join, outer_join]).filled(999.)
```
profxj, line 479: same as above

profxj (changes requested):
Thanks!

My only concern related to the 999 flags

If we are sure removing those is fine, then go for it.

aelanman (approved):
Sunil asked me to look at the Surveys.ipynb notebook; That part looks fine to me! I don't see anything that would break my code.

PR #239, opened 2026-05-18
DL Querry Client Error work around

lordrick94:


lordrick94:
Apologies, wrong commit message, this is a work around for the recurring query client error on datalab, for which one can quickly independently check if coord is in DECaL footprint

frb/surveys/survey_utils.py
```diff
@@ -136,6 +137,31 @@ def is_inside(surveyname:str, coord:SkyCoord)->bool:
...
+        print(f"\n{'='*60}")
+        print(f"ERROR: {err_type} while querying {surveyname}")
+        print(f"Coordinate: RA={coord.ra.deg:.6f}, Dec={coord.dec.deg:.6f}")
+        print(f"Error details: {str(e)[:200]}")
+        print(f"{'='*60}")
+        print(f"\nPlease verify manually if this coordinate is inside {surveyname}.")
+        print("You can check at: https://www.legacysurvey.org/viewer")
+
+        while True:
+            user_input = input(f"Is coordinate inside {surveyname}? (true/false): ").strip().lower()
```
SunilSimha, line 158: Perhaps have an option to let the user skip if legacysurvey.org is also down simultaneously.

SunilSimha (changes requested):
Thanks for this. Though I must ask, does it happen often that datalab is down but legacysurvey.org is up? I've generally had trouble with IRSA in the past but rarely datalab.

Requesting a minor change.

